PR #22, opened 2015-05-26
fixes #21: use setImmediate in _setTimeout when millis is zero

### LINKED ISSUE #21: use setImmediate in _setTimeout when millis is less than 4
The HTML5 `setTimeout` function will use a value of four for any timeout value less than four (see [step 7 in Section 8.4](https://html.spec.whatwg.org/multipage/webappapis.html#timers)). `_setTimeout` should use [`setImmediate`](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate) for calls with a `millis` value less than four, or at least for calls with a `millis` value of zero.

michaelficarra:
Fixes #21. I used the conservative implementation, choosing to only use `setImmediate` when `millis` is zero or lower. I figure that a value of `1` is much closer to `4` than it is to `0` given a sufficiently fast computer.

### DIFF
--- a/MODULES.md
+++ b/MODULES.md
@@ -108,6 +108,14 @@ later' :: forall e a. Number -> Aff e a -> Aff e a
 Runs the specified asynchronous computation later, by the specified
 number of milliseconds.
 
+#### `finally`
+
+``` purescript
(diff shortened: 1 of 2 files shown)


jdegoes:
Awesome. :+1:

PR #24, opened 2015-06-12
Updates for psc-0.7

anttih:
This is a work in progress. It compiles, but only with the latest pulp. Need to wait for the grunt plugin to get updated and requires more testing.


jdegoes:
Thanks, @anttih. Happy to keep this open till it's ready to merge and 0.7 comes out.

garyb:
There's a version of the gulp plugin that works with the new compiler `0.5.0-rc.1`. I think you have to specify that as the lower bound for it to pick it up though, it won't match with `^0.5.0`.

anttih:
Should `Control.Monad.Aff.Debug.Trace` be renamed to `Control.Monad.Aff.Console`? We could also rename `trace` in that module to `print`, to be in line with the Console module.

garyb:
Yep, sounds good. :+1: Maybe add `log` in there while you're at it too.

anttih:
This is looking good now. I replaced grunt with gulp but kept the build commands the same. The docs are now generated into separate files and linked to from the README.

anttih:
Just noticed the `gulp psc` command is not generating the js bundle.

jdegoes:
@anttih Can you squash commits so we get a clean commit history? Thanks!

anttih:
Done. We'll need to wait for gulp-purescript in order for the build to work again, however. :(

anttih:
@jdegoes @garyb Updated to the new gulp-purescript, and the build is working again.

Travis is running the build but we don't have .travis.yml. Should I add that?

garyb:
Yes please, it's probably supposed to have one but we forgot to add a `!/.travis.yml` to the `.gitignore` file.

jdegoes:
Thanks for your work on this @anttih! :+1:

PR #26, opened 2015-07-16
Add MonadRec instance and test

garyb:
Moved things around a bit in the build so the examples files aren't generated in `output`, as that was interfering with bundling.


jdegoes:
:+1: Nice.

PR #27, opened 2015-07-17
Add MonadAff instances for transformers

garyb:


garyb:
Do you think it's worth providing a `MonadCont (Aff e)` instance too, or is that problematic with Aff's error handling? I guess it should be fine, as you can't raise an error when resuming the continuation.

jdegoes:
:+1:

jdegoes:
> Do you think it's worth providing a MonadCont (Aff e) instance too, or is that problematic with Aff's error handling?

I don't think it's a problem. You can try it and find out. :smile:

Issue #32, opened 2015-08-15
Is launchAff too dangerous?


hdgarrood:
Currently, `launchAff` throws away all errors. This might not be so bad in a browser, as the errors might appear in the console anyway? It's particularly bad on node, though, as there's no indication at all that an error has occurred.

I am aware that this is intended and documented behaviour, but I think it's still very easy to be tripped up by this. I think there's a case to be made that it's too dangerous to be exposed as is, especially since if users really want that functionality, they can easily implement it themselves in terms of `runAff`.

What do you think about, say, `launchAff = runAff throwException (const (pure unit))` instead?

beckyconning:
:+1:

jdegoes:
:+1: 

Another option would be using the global error handler, although I don't know if that's portable to node. But I think I like just throwing the exception better.

hdgarrood:
I just tried this and it's a little harder than I thought to make it typecheck, because `runAff` forces you to have the same row of effects in the `Eff` you get out. I want to write this:

```purescript
  -- | Converts the asynchronous computation into a synchronous one. All values
  -- | are ignored, and if the computation produces an error, it is thrown.
  launchAff :: forall e a. Aff e a -> Eff (err :: EXCEPTION | e) Unit
  launchAff = runAff throwException (const (pure unit)) <<< liftEx
    where
    liftEx :: forall e a. Aff e a -> Aff (err :: EXCEPTION | e) a
    liftEx x = x
```

But `liftEx` doesn't typecheck. Any suggestions?

paf31:
`unsafeInterleaveEff`?

garyb:
If you defined it as 
``` purescript
launchAff :: forall e a. Aff (err :: EXCEPTION | e) a -> Eff (err :: EXCEPTION | e) Unit
```

That would work too right? You don't _have_ to use the effect just because it's there.

hdgarrood:
Yes, but I think that would prevent you from using a closed row of effects for an `Aff` action unless that row contained `err :: EXCEPTION`, which doesn't seem right, as it's an implementation detail, and since `Aff` already includes its own error handling.

I guess `unsafeInterleaveEff` works. Is there no "above-board" way of doing this?

beckyconning:
Are we dead set on keeping launchAff?

garyb:
For the closed row case I guess you could go via `runAff`?

hdgarrood:
> Are we dead set on keeping launchAff?

[...]

PR #33, opened 2015-08-16
Stop ignoring exceptions in `launchAff`

hdgarrood:
`launchAff` will now throw exceptions rather than swallowing them. Fixes
issue #32

### DIFF
--- a/src/Control/Monad/Aff.purs
+++ b/src/Control/Monad/Aff.purs
@@ -26,7 +26,7 @@ module Control.Monad.Aff
   import Control.Monad.Cont.Class(MonadCont)
   import Control.Monad.Eff(Eff())
   import Control.Monad.Eff.Class(MonadEff, liftEff)
-  import Control.Monad.Eff.Exception(Error(), EXCEPTION(), catchException, error)
+  import Control.Monad.Eff.Exception(Error(), EXCEPTION(), catchException, throwException, error)
   import Control.Monad.Eff.Unsafe(unsafeInterleaveEff)
   import Control.Monad.Error.Class(MonadError, throwError)
   import Control.Monad.Rec.Class(MonadRec, tailRecM)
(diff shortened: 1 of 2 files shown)


src/Control/Monad/Aff.purs
```diff
@@ -65,9 +65,12 @@ module Control.Monad.Aff
   cancelWith aff c = runFn3 _cancelWith nonCanceler aff c
 
   -- | Converts the asynchronous computation into a synchronous one. All values
-  -- | and errors are ignored.
-  launchAff :: forall e a. Aff e a -> Eff e Unit
-  launchAff = runAff (const (pure unit)) (const (pure unit))
+  -- | are ignored, and if the computation produces an error, it is thrown.
+  launchAff :: forall e a. Aff e a -> Eff (err :: EXCEPTION | e) Unit
```
jdegoes: The `EXCEPTION` may be thrown asynchronously, whereas the current signature implies that the returned `Eff` will throw the exception -- and therefore, that such exceptions can be caught using `catchException`. The implication is invalid in the general case, only valid for purely synchronous `Aff`. 

Maybe it suffices to amend the comment to, "... and if the computation produces an error, it is thrown, but not necessarily in the current thread of execution.".

jdegoes: Originally `Aff` had something like a `LATER` effect type which could itself accept a row of effects, but that turned out to have problems with type inference. As a consequence, distinctions between "effects now" and "effects later" can't be captured.

hdgarrood: Ah, of course, thanks. Shall I just amend the comment then, or do you want to see if you can think of a better way of writing it?

jdegoes: The only other alternative I can see is not adding the `EXCEPTION` effect type, so as not to give the user the (false) assurance they can catch it. It's fundamentally un-catchable. Maybe that's a sign the user should always have to supply an error handler (and we could provide a global one if it makes sense?).

hdgarrood: Ok then - so would we end up with something like `launchAff = flip runAff (const (pure unit))` then? What would be the type of the global error handler we might provide?

hdgarrood: Or perhaps we could leave the definition like this, but just change the returned type, to, say `Eff (errUncatchable :: EXCEPTION | e) Unit`?

jdegoes: After mulling this over the past couple days, I think the best option right now is to leave it as is, but add the comment. This way you can still 'catch' a synchronous exception (which may in fact happen, since `Aff` is not *necessarily* asynchronous), while the comment will warn you away from assuming that future exceptions can be caught synchronously.

Now what we'd really like is for all `main` functions to expect `Aff` and not `Eff`. :wink:

hdgarrood:
Ok, sounds good - I added a comment about asynchronousness and catching exceptions. Although, it has just occurred to me that JavaScript doesn't really have "threads", does it?

hdgarrood:
Ok, how's this?

jdegoes:
Thanks!

PR #35, opened 2015-08-26
Fix link in README.

fisx:
not sure if this is what you want, but it should make the link work under http://pursuit.purescript.org/packages/purescript-aff/0.11.0


jdegoes:
:+1: Thanks!

PR #36, opened 2015-08-26
Auto-deploy to pursuit on tagged release

garyb:
I can't tell for sure if this will work, so if it doesn't I may end up making changes and creating/deleting a tag a few times to test. Once we figure it out on here it can be applied everywhere at least!

I've setup the `$PURSUIT_TOKEN` variable in travis already.


garyb:
I couldn't get it to do even run `psc-publish` so I'll try and figure that out on a temp repo somewhere then come back to this.

garyb:
I reverted also reverted all the commits I made while trying to get it to work.

PR #45, opened 2016-02-13
Add forkAll combinator

natefaubion:


docs/Control.Monad.Aff.md
```diff
@@ -159,6 +159,16 @@ will not block on the result of the computation.
 Returns a canceler that can be used to attempt cancellation of the
 forked computation.
 
+#### `forkAll`
+
+``` purescript
+forkAll :: forall f e a. (Foldable f) => f (Aff e a) -> Aff e Unit
+```
+
+Forks many asynchronous computation at once, ignoring the results.
```
jdegoes: It would be good to add a guarantee that all computations will be forked synchronously.

src/Control/Monad/Aff.js
```diff
@@ -96,6 +96,26 @@ exports._forkAff = function (nonCanceler, aff) {
...
+      };
+    })({})(affs);
+
+    try {
+      success({});
+    } catch(e) {
+      error(e);
+    }
+
+    return nonCanceler;
```
jdegoes: It would be nice to be able to cancel these by propagating the canceling behavior into the forked computations, rather than returning a non-canceler here.

natefaubion: Ok, so to make sure I'm getting it right: Collect all the cancellers in the fold, then return a canceller that propagates to all those cancellers. Should I also push that array of cancellers to `success`?

jdegoes: A canceller is, I believe, an `Aff` which contains a function which takes an error and returns an `Aff` of a boolean. So yes, we'd collect all the cancellers in the fold, and then do some really messy logic (in JS, anyway) to propagate the cancel signal to the cancellers, and then back-propagate the booleans from the sub-cancellers to the parent canceller. I think you can find examples of this elsewhere in the JS code.

natefaubion: I get what the canceller does, and yes, it's very similar to what `cancelWith` does. Should this be part of the primary canceller for this action, or should it push this canceller to `success` like in `forkAff`?

jdegoes: Oh, cancelers have a monoid, right? So you could just pass append in and smash them all together. That'd be easier.

natefaubion: It does, which is very convenient, but doesn't preserve the stack-safe property since it uses the Applicative instance for Aff.

natefaubion: I guess you're only going to need cancelers with truly async computations, so maybe its not a big deal?

jdegoes: Agreed.

natefaubion:
I went ahead and made it mirror `forkAff`, so that it returns a canceler that can be used to cancel all of the forked computations. I also went ahead and made the canceler stack safe, since it was very easy to trigger a stack overflow using the Monoid instance (I could only run the test I made with a small number since the timeout canceler is synchronous).

jdegoes:
@natefaubion :+1: Looks great, @garyb you want to do a final review?

garyb:
Will do, once 2.5 is out!

src/Control/Monad/Aff.js
```diff
@@ -96,6 +96,61 @@ exports._forkAff = function (nonCanceler, aff) {
...
+exports._forkAll = function (nonCanceler, foldl, affs) {
+  var voidF = function(){};
+
+  return function(success, error) {
+    var cancelers = foldl(function(acc) {
+      return function(aff) {
+        acc.push(aff(voidF, voidF));
+        return acc;
+      }
+    })([])(affs);
```
garyb: :see_no_evil:

garyb:
Looks good to me too! :+1:

src/Control/Monad/Aff.js
natefaubion: Hah. I don't love it either :/

I might be able to extract out a helper to share with cancelWith.

garyb: Nah, I can see it's fine, it's just the mutating of the array in the fold while constructing it that made me look twice. :smile:

natefaubion: Ah. Thought you were referring to the sheer length of this function. Hard to tell in email sometimes

PR #47, opened 2016-02-15
Fix warnings generated by purescript-0.8.0

telser:


jdegoes:
@telser Nice work! The Travis CI build failed. Can you look into it? Thanks!

telser:
@jdegoes Thanks! It seems that purescript-0.8 needs node>=4, but this has specified node 0.12 in the travis.ci config. If there is a preference for the node version to use here, I can make that change and squash back down to one commit.

garyb:
That's just leftover stuff, most of the travis scripts I've been adding/updating lately I've switched it to use 5.

I'd suggest holding off merging this until we release PS 0.8.1, as after that is released we're going to update and major-bump all the core libraries (updating for 0.8-without-warnings requires breaking changes), so if we release this now for we'll have to release another major-bumped version soon anyway.

telser:
@garyb I updated the pr to use node "4" since according to https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Provided-Node.js-Versions at least, 5 isn't supported yet. Maybe 5 is if you pay or is planned, not sure, but this at least gets the build to pass. If there is anything else I can do here, or on some other pr I've submitted, just let me know!

PR #48, opened 2016-03-04
Removing shadowing in catch blocks.

lunaris:
The current codebase uses `e` as the standard name given to exceptions caught by `catch` blocks. In many cases this shadows an existing `e` in scope. This is fine in most scenarios, but it trips up some minification and uglification processes (see for instance https://github.com/mishoo/UglifyJS2/issues/409#issuecomment-149593800). Since it does no harm to rename such variables to avoid the shadowing in the first place, this PR performs such a renaming to (hopefully) enable more minification tools to work properly without changing the library's semantics.

### DIFF
--- a/src/Control/Monad/Aff/AVar.js
+++ b/src/Control/Monad/Aff/AVar.js
@@ -9,10 +9,10 @@ exports._makeVar = function (nonCanceler) {
       success({
         consumers: [],
         producers: [],
-        error: undefined 
+        error: undefined
       });
-    } catch (e) {
-      error(e);
+    } catch (err) {
+      error(err);
     }
 
     return nonCanceler;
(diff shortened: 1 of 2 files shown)


jdegoes:
@lunaris Would you mind changing it from `x` to `err`? I feel `err` conveys more information than `x`.

lunaris:
@jdegoes Sure; it's now `err`. I picked `x` initially because it is used in some other languages (as in eXception) and because it's not used anywhere else in the foreign JS code. `err` is used in a couple of other places but I'm pretty sure it's not shadowing so should all be good. Hopefully this is good to go.

lunaris:
@jdegoes Is this good to go now? Looks to be OK with Travis.

src/Control/Monad/Aff/AVar.js
```diff
@@ -81,8 +81,8 @@ exports._killVar = function (nonCanceler, avar, e) {
 
         try {
           consumer.error(e);
```
jdegoes: There's a reference to `e` here, is this correct or should this have been updated?

jdegoes:
:+1:

PR #49, opened 2016-03-08
Add `Affable` class

garyb:
I didn't add the various division between `Affable` and `MonadAff` as I guess we should come up with laws for them, but I've not figured out what they should be yet as I'm shaky on the monad morphism stuff even.

This is a breaking change so will need a version bump again.


jonsterling:
When I see this, I can't help but think of the classic Larry David / Curb Your Enthusiasm line, “I was just trying to be `Affable`!”

garyb:
I guess `FunctorAff`'s law would be that it forms a natural transformation of the proper `forall f g a. (Functor f, Functor g) => f a -> g a` type (we usually ignore the functor constraint).

garyb:
@jonsterling I don't remember who suggested `Affable`, I thought it might have been you actually, but it's genius anyway :smile:. `Effable` likewise if/when we introduce it.

jdegoes:
:+1:

jdegoes:
Although I am curious why `fromAff` and not `liftAff`.

Also, would it be a good idea to imply `MonadAff` from any `Monad` which is `Affable`? This would basically ensure no one can implement `MonadAff` directly, rather, any `Monad` which is `Affable` automatically gets a `MonadAff`.

garyb:
I wasn't sure if we wanted to keep the same function to work for both, that's fine by me too. I figured `liftAff` implies something more lawful than just `toAff`?

For the latter thing, do you mean including an instance like this?
``` purescript
instance impliedMonadAff :: (Monad m, Affable e m) => MonadAff e m where
  liftAff = fromAff
```

I'm actually not sure if that works, I can give it a try, but also I thought tit's maybe possible to implement an instance where the `Affable` for a `Monad` without making it a monad morphism? If we're just using `MonadAff` to imply both constraints it seems a bit unnecessary and we could just use both constraints at the definition site instead.

jdegoes:
OK, good to ship as is (I still go back and forth on the name, but I'm good with trying this).

garyb:
I went back to just providing `liftAff` under `Affable` and not having a separate function for `MonadAff`, I think it will be less confusing this way. Added some minimal comments to the meaning of the classes too, but I'm not exactly sure how to state the mmorph laws for `MonadAff` so just left the specifics unstated.

garyb:
I realised that a fairly important thing was missing: an instance for `Free`, which was pretty much the motivating case in the first place. I had to add the dependency for that to make it possible, but I think it's worth it.

I also updated the build to use `pulp`, added auto-publish to pursuit, fixed the license description in `bower.json`, added a dependency status badge, and moved the "examples" into the test suite which will ensure they run as part of Travis.

jdegoes:
👍

paf31:
> A class for types that can carry an `Aff` value by some means.

Is the goal here to use `MonadAff` with `Free f` when `f` is `Affable` but not a `Monad`?

What are the laws for `Affable`? If it's that `liftAff` is a monad morphism, then is something like `liftAff _ = throwError "explode"` a valid instance, for example? Or is it that `liftAff` identifies `Aff` as a a summand of `m` as a coproduct? In which case the instances for the transformers seem off.

garyb:
Affable is lawless, it's basically a specialised `inject`. When it is also `MoandAff` then it's a monad morphism.

jdegoes:
I should think the laws could be something like:

```purescript
liftF (liftAff (pure a)) = pure a
liftF (liftAff (aff >>= f)) = liftF (liftAff aff) >>= (liftF (f >>> liftAff))
```

paf31:
So the laws are relative to `Free`? That seems a bit specialized to me. `Aff` doesn't really have anything to do with `Free` in general.

It might also be interesting to try coming up with a class with the `Aff` operations, rather than lifting everything at once. Then you could talk about morphisms between `Affable` things using the type class morphisms approach. I dunno.

jdegoes:
> So the laws are relative to Free? That seems a bit specialized to me. Aff doesn't really have anything to do with Free in general.

It's true, we could put this into `purescript-aff-free` or something. The question is, do the above `Free`-based laws imply the monad transformer laws when `m` is a monad? That might be an argument for `Affable` being a superclass of `MonadAff` (though I need to think about it more).

> It might also be interesting to try coming up with a class with the Aff operations, rather than lifting everything at once.

Yeah, I like that approach. Though I wonder what laws one could state with such an approach, without assuming that `f` is a monad...

paf31:
> It's true, we could put this into `purescript-aff-free` or something.

Perhaps that would be a good approach. My only concern is that `aff` is a core part of the "real world PureScript" ecosystem, whether or not it's part of the core/contrib libraries. It's really useful for beginners who want to create their first real project, and it would be a shame to make the API more confusing with something which seems primarily motivated by a need in Halogen-style apps.

jdegoes:
> Perhaps that would be a good approach.

I'm good with that. @garyb? The main thing it changes is that `Affable` is not a superclass of `MonadAff`. However, a `MonadAff` could imply `Affable`, and we could go back to the old `fromAff` name.

garyb:
That's fine by me. Do we need a separate library, I could just split `Affable` out within `purescript-aff`?

I'm not too sure about the free-based laws holding the way we use it either, as we can have nested `Free`s, the `Aff` inside a `Coproduct`, etc.

jdegoes:
> Do we need a separate library, I could just split Affable out within purescript-aff?

I think that's fine for now (in its own separate module, separate doc, etc.). We can move it to a new library if it seems the approach is working and generally useful (and do the same for `Eff`).

> I'm not too sure about the free-based laws holding the way we use it either, as we can have nested Frees, the Aff inside a Coproduct, etc.

The laws should hold no matter the nesting, so long as they hold individually at each level. Right?

PR #50, opened 2016-03-10
Affable, take two

garyb:
- `Affable` is now separate from `MonadAff`
- There's an `instance affableMonadAff :: (MonadAff eff m) => Affable eff m`. I'm not sure how useful it will be currently, it might need fundeps to behave well without use of type annotations
- I added `MonadEff` as a superclass of `MonadAff` - I think that should always have been the case if we're saying `Aff` is the moral equivalent of `ErrorT (ContT Unit (Eff e) a`?

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -19,7 +19,6 @@
   "dependencies": {
     "purescript-console": "^0.1.0",
     "purescript-exceptions": "^0.3.0",
-    "purescript-free": "^0.9.1",
     "purescript-functions": "^0.1.0",
     "purescript-transformers": "^0.8.1"
   }
--- a/src/Control/Monad/Aff/Class.purs
+++ b/src/Control/Monad/Aff/Class.purs
@@ -4,8 +4,8 @@ import Prelude
 
 import Control.Monad.Aff (Aff())
 import Control.Monad.Cont.Trans (ContT())
+import Control.Monad.Eff.Class (MonadEff)
 import Control.Monad.Except.Trans (ExceptT())
-import Control.Monad.Free (Free(), liftF)
 import Control.Monad.List.Trans (ListT())
 import Control.Monad.Maybe.Trans (MaybeT())
 import Control.Monad.Reader.Trans (ReaderT())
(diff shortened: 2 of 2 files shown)


jdegoes:
:+1: Unless @paf31 has any comments.

garyb:
I'll wait for Phil to chime in before releasing this time :smile:

paf31:
I'd just like to understand what the difference is between the two classes. I always thought of `MonadAff` as meaning "this monad is isomorphic to a monad transformer with `Aff` at the bottom". So maybe you could assume something along the lines of `liftAff = lift <<< ... <<< lift` (not exact) and reason about the lifted operations in `Aff` that way.

But then, what is `Affable`? It seems to allow `Aff` to appear in a summand of the base functor of `Free`, _or any of the cases listed above_. In which case, how do you characterize it? What about something like `FreeT` - if that is allowed an instance one day then which instance does `FreeT _ (Aff _)` get? What's an example of usage of each type of instance?

Sorry if it seems like I'm butting in here, I know this is a company repo, it's just that `aff` is special somehow, since it's used so widely.

garyb:
Yeah, that's the way I see `MonadAff` too pretty much. `FreeT` would be `MonadAff` I'd expect, being that it's a transformer.

`Affable` is a case of "can eventually be interpreted as `Aff`" I guess, and because of that I'm not really sure if you can formulate sensible laws to describe the behaviour since it depends on the interpreter as much as the type.

I understand where you're coming from with these questions, and in theory `Affable` should be redundant as we should be interpreting some `Free` into `Aff` anyway, but that isn't super practical, sometimes you want something which is part-`Aff` and part-something-that-will-become-`Aff`.

We basically already have this locally in Halogen, maybe we should just keep it there and then it's less of a sticking point, as it's a thing for Halogen's convenience specifically, which realistically is probably where it is most useful anyway.

garyb:
@jdegoes I spoke to Phil on IRC again yesterday and perhaps for now the least controversial approach would be to not provide `Affable` as part of `aff`, as it gives it to much prominence when we haven't exactly nailed down a proper description for it, and it's mostly useful for Halogen.

I'll move it into Halogen for now, to replace `FunctorAff` we have in there currently that is basically the same thing.

jdegoes:
@garyb Sounds good. I'd also be in favor of a `aff-free` library, since I do believe `Affable` has laws that can be stated with `Free`.

garyb:
Ah yes, I'd forgotten about that version. I'll perhaps put that together then - I suppose we omit the instances for transformers with that formulation, but I don't think that's an issue as I don't think they're used anyway.

PR #52, opened 2016-03-23
Allow cancellation of computations forked with `runAff`

sardonicpresence:
At present, in order to have a cancellable asynchronous computation it seems you forego both success and error callbacks in the call to `forkAff`, the only function to return a `Canceler`.

Particularly given that in the JavaScript `runAff` already returns a `Canceler` I saw no reason not to include a version returning it in PureScript.

In the process of adding a test for the above, I was inconvenienced by the inability to
```
do
  c <- runAff throwException (const (pure unit)) a
  launchAff $ cancel c (error msg)
```
due to the odd coersion of effect records in `launchAff` so I thought I'd propose a change there as well.

I'm assuming that the breaking nature of these changes will be contentious at least but thought I'd start here.

### DIFF
--- a/src/Control/Monad/Aff.purs
+++ b/src/Control/Monad/Aff.purs
@@ -52,7 +52,7 @@ foreign import data Aff :: # ! -> * -> *
 -- | asynchronous computation.
 type PureAff a = forall e. Aff e a
 
--- | A canceler is asynchronous function that can be used to attempt the
+-- | A canceler is an asynchronous function that can be used to attempt the
 -- | cancelation of a computation. Returns a boolean flag indicating whether
 -- | or not the cancellation was successful. Many computations may be composite,
 -- | in such cases the flag indicates whether any part of the computation was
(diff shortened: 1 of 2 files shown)


src/Control/Monad/Aff.purs
```diff
@@ -75,17 +76,22 @@ cancelWith aff c = runFn3 _cancelWith nonCanceler aff c
 -- | If you do need to handle exceptions, you can use `runAff` instead, or
 -- | you can handle the exception within the Aff computation, using
 -- | `catchError` (or any of the other mechanisms).
-launchAff :: forall e a. Aff e a -> Eff (err :: EXCEPTION | e) Unit
-launchAff = runAff throwException (const (pure unit)) <<< liftEx
-  where
-  liftEx :: Aff e a -> Aff (err :: EXCEPTION | e) a
-  liftEx = _unsafeInterleaveAff
+launchAff :: forall e a. Aff (err :: EXCEPTION | e) a -> Eff (err :: EXCEPTION | e) Unit
```
jdegoes: Why not modify this to return a canceller, to? User can always ignore a canceler if it's not necessary. Also this would allow you to delete the helper function `runAff_`.

jdegoes:
@sardonicpresence Howdy there, waiting for your feedback above. Thanks for the PR, by the way!

src/Control/Monad/Aff.purs
sardonicpresence: No reason other than to minimise changes to the API. I certainly don't object to your suggestion.
Thanks for taking a look btw.

jdegoes: Since this will be included in a breaking release anyway (due to 0.9 updates), it's a good chance to clean up things like this. I'd rather a user `void` a return value they don't want, then not have the option (or multiply the surface area of the API).

sardonicpresence:
Actioned review comments and re-based (hope that's reasonable etiquette).

src/Control/Monad/Aff.purs
```diff
@@ -73,15 +73,15 @@ cancelWith aff c = runFn3 _cancelWith nonCanceler aff c
 -- | If you do need to handle exceptions, you can use `runAff` instead, or
 -- | you can handle the exception within the Aff computation, using
 -- | `catchError` (or any of the other mechanisms).
-launchAff :: forall e a. Aff e a -> Eff (err :: EXCEPTION | e) Unit
-launchAff = runAff throwException (const (pure unit)) <<< liftEx
-  where
-  liftEx :: Aff e a -> Aff (err :: EXCEPTION | e) a
-  liftEx = _unsafeInterleaveAff
+launchAff :: forall e a. Aff (err :: EXCEPTION | e) a -> Eff (err :: EXCEPTION | e) (Canceler (err :: EXCEPTION | e))
```
jdegoes: These changes look great. This is the only one I'm concerned about. Technically, an `EXCEPTION` effect type in an `Aff` doesn't mean anything, since `Aff` doesn't throw exceptions in a way that can be caught via the same mechanism for `Eff`.

The intention is that you can lift any `Eff` with `EXCEPTION` into an `Aff`, and it drops the `EXCEPTION`; and that you can lower any `Aff` into an `Eff` and it adds `EXCEPTION`. That's not quite precise but I think it's probably the best we can do.

Is there some other way of addressing your need for this change?

sardonicpresence: To be frank I really struggle to see the utility of the effect record mechanism in Purescript generally.

I understand what you saying in regards to `EXCEPTION` not making sense in `Aff`, in fact this is the first instance I think I've encountered where rejecting certain effects makes sense like this. However `runAff` would still allow you to run an `Aff` with an `EXCEPTION` effect for example.

I think what you actually want is probably `runAff` to be able to declare that the effects in the `Eff` monads are a superset of those in the `Aff` monad but I don't believe that's currently possible in the language. I'm not sure what would be correct for the `Canceler`.

I hadn't thought of `liftEff'` as a kind of dual for `launchAff`; I'll have a think as to whether I can make use of it though I'm concerned by the use of `Either` as opposed to calling the error callback at a glance.

I'll have another stab at it tonight.

sardonicpresence:
Alright I reverted the objectionable change, requiring further coercing in `launchAff`. If we could be more expressive about relationships between effect records I'd do differently but there we are.

I've also re-based once again.

jdegoes:
@sardonicpresence Great work. Thanks! 🙏

PR #56, opened 2016-05-23
Propagate Par errors

spencerjanssen:
I noticed that failing `Par` computations don't pass the exception on, instead they silently lose the exception.


jdegoes:
@spencerjanssen Nice work! Thank you. 🙏

PR #57, opened 2016-06-01
Alternative tailRecM implementation

natefaubion:
The current implementation of `tailRecM` is problematic because it will asynchronously bounce completely synchronous computations. This breaks the guarantee Aff provides that synchronous computations will always execute synchronously. In fact, one could argue that the current implementation breaks equational reasoning by introducing an observable difference between (essentially) `leftEff do a; b` and `do liftEff a; liftEff b`.

This implementation is a bit hairier, but by observing synchronous effects in `tailRecM`, we can guarantee that a totally synchronous computation remains synchronous, and that it is stack safe.

This fixes SD-1665, and all tests pass on SlamData. Thoughts on this implementation?

### DIFF
--- a/src/Control/Monad/Aff.js
+++ b/src/Control/Monad/Aff.js
@@ -293,3 +293,64 @@ exports._liftEff = function (nonCanceler, e) {
     return nonCanceler;
   };
 }
+
+exports._tailRecM = function (isLeft, f, a) {
+  return function(success, error) {
+    return function go(acc) {
+      var result, status, canceler;
+
+      // Observes synchronous effects using a flag.
+      //   status = 0 (unresolved status)
+      //   status = 1 (synchronous effect)
+      //   status = 2 (asynchronous effect)
+      while (true) {
+        status = 0;
+        canceler = f(acc)(function(v) {
+          // If the status is still unresolved, we have observed a
+          // synchronous effect. Otherwise, the status will be `2`.
+          if (status === 0) {
+            // Store the result for further synchronous processing.
+            result = v;
+            status = 1;
+          } else {
+            // When we have observed an asynchronous effect, we use normal
+            // recursion. This is safe because we will be on a new stack.
+            if (isLeft(v)) {
+              go(v.value0);
+            } else {
+              try {
+                success(v.value0);
+              } catch (err) {
+                error(err);
+              }
+            }
+          }
+        }, error);
+
+        // If the status has already resolved to `1` by our Aff handler, then
+        // we have observed a synchronous effect. Otherwise it will still be
+        // `0`.
+        if (status === 1) {
+          // When we have observed a synchronous effect, we merely swap out the
+          // accumulator and continue the loop, preserving stack.
+          if (isLeft(result)) {
+            acc = result.value0;
+            continue;
+          } else {
+            try {
+              success(result.value0);
+            } catch (err) {
+              error(err);
+            }
+          }
+        } else {
+          // If the status has not resolved yet, then we have observed an
+          // asynchronous effect.
+          status = 2;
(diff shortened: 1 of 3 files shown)


garyb:
I'm about to go out so don't have time to review this right now, but I found SD-1665 horrifying when I read about it this morning, so I'm glad it seems like there might be a reasonable way out of it.

natefaubion:
I have pushed some inline comments to hopefully better explain what is going on.

src/Control/Monad/Aff.js
```diff
@@ -293,3 +293,70 @@ exports._liftEff = function (nonCanceler, e) {
...
+      var result, status, canceler;
+
+      // Observes synchronous effects using a flag.
+      //
+      //   status = 0 (unresolved status)
+      //   status = 1 (synchronous effect)
+      //   status = 2 (asynchronous effect)
+      //
+      // The inner Aff is run, and if the status is still 0, we have observed
+      // a synchronous effect. The result is stashed for later synchronous
```
jonsterling: Can you explain why the status being still `0` means we've observed a synchronous effect? I'm sure you're right, but since I don't really know the guts of Aff too well, it wasn't quite clear to me how.

src/Control/Monad/Aff.js
```diff
@@ -293,3 +293,70 @@ exports._liftEff = function (nonCanceler, e) {
...
+      while (1) {
+        status = 0;
+        canceler = f(acc)(function(v) {
+          if (status === 0) {
+            result = v;
+            status = 1;
+          } else {
+            try {
+              if (isLeft(v)) {
+                go(v.value0);
```
jdegoes: This recursion looks stack unsafe.

jdegoes:
👍 To this way of solving it.

src/Control/Monad/Aff.js
natefaubion: The recursion is in the "asynchronous" branch. It only is triggered when we have observed an asynchronous effect, and we are thus on a new stack.

src/Control/Monad/Aff.js
natefaubion: The order of synchronous effects are:

* Initialize status to 0
* Run Aff
* Observe status

If the status changed between initializing and observing it, our Aff has resolved synchronously. When resolving synchronously, we check that the status has NOT changed from its initial state. You have to observe the change in two places: during resolution and, after launching the Aff.

natefaubion:
Also if anyone has any ideas to test any edge cases, I'm happy to add it.

src/Control/Monad/Aff.js
jonsterling: OK, thanks! I think it makes sense to me what is going on here now.

natefaubion:
I've pushed (what I think are) better inline comments, and also added a test that uses `tailRecM` to interleave synchronous and asynchronous effects.

natefaubion:
I have added a test that verifies that a completely synchronous `tailRecM` is indeed resolved synchronously.

jdegoes:
:shipit: Nice work @natefaubion!

jonsterling:
Great job figuring this out, @natefaubion!

ajnsit:
> the guarantee Aff provides that synchronous computations will always execute synchronously

Ah, this helps explain a bug I was seeing, where forking an aff would still call the handler synchronously. Is this documented properly somewhere? Also, is there a way to force a true fork irrespective of whether the computation is synchronous or asynchronous.

safareli:
`delay` with 0 or some small number might do the trick
``` purs
fork do
 delay (Milliseconds 0.0)
 ..
```

PR #68, opened 2016-08-07
Add `peekVar` for reading from an avar without consuming

garyb:
We sometimes run into cases where an `AVar` behaving like a `Ref` would be useful. Having a `peekVar` operation makes that case less error prone than the current situation of having to remember to put the value back after reading it every time.

### DIFF
--- a/src/Control/Monad/Aff/AVar.purs
+++ b/src/Control/Monad/Aff/AVar.purs
@@ -5,6 +5,7 @@ module Control.Monad.Aff.AVar
   , makeVar
   , makeVar'
   , takeVar
+  , peekVar
   , putVar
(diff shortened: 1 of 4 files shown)


jdegoes:
Looks good, though perhaps better as low-level JS operation so it's not dependent on the current non-blocking semantics of put.

garyb:
Good thinking :+1:

natefaubion:
Agreed with @jdegoes, I'd like this to truly not consume anything. This implementation will just put it back onto the tail of the producer queue, which may have other things queued ahead of it.

garyb:
I've reworked this now to avoid consuming at all, and added some test cases.

PR #75, opened 2016-10-21
Remove MonadCont instance

### LINKED ISSUE #46: `monadCont` instance doesn't preserve cancellation behavior
I found this while working on a test for cancelling `forkAll`.
```purescript
delay :: forall eff. Int -> Aff eff Unit
delay n = callCC \cont ->
  later' n (cont unit)

fork :: forall eff. Aff (console :: CONSOLE | eff) Unit
fork = do
  a <- forkAff (later' 10000 (Console.log "oops"))
  cancel a (error "bye")
  Console.log "Canceled a"
  b <- forkAff (do delay 10000
                   Console.log "oops")
  cancel b (error "bye")
  Console.log "Canceled b"
```

I would assume that the two formulations would be identical, but the `callCC` version does not cancel correctly. The first variation (using `later'` directly) gets cancelled correctly and prints "Canceled a". While the `callCC` variation waits 10 seconds, _then_ get's cancelled, but does not print _anything_.

### LINKED ISSUE #59: Remove `MonadCont` instance
Since we don't want to encourage continuation-style behaviour for Aff

garyb:
Resolves #59, also resolves #46 since if there's no instance it can't be broken :smile:

@jdegoes are you happy with this now? At the time I opened #59 you said to keep it around for a while longer.

### DIFF
--- a/src/Control/Monad/Aff.purs
+++ b/src/Control/Monad/Aff.purs
@@ -25,7 +25,6 @@ import Prelude
 import Control.Alt (class Alt)
 import Control.Alternative (class Alternative)
 import Control.Monad.Aff.Internal (AVBox, AVar, _killVar, _putVar, _takeVar, _makeVar)
-import Control.Monad.Cont.Class (class MonadCont)
 import Control.Monad.Eff (Eff)
 import Control.Monad.Eff.Class (class MonadEff)
 import Control.Monad.Eff.Exception (Error, EXCEPTION, throwException, error)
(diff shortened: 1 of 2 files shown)


jdegoes:
Yes, let's do it! BREAK ALL THE THINGZ!!!

garyb:
I'm not sure if it will actually break anyone's code anyway, the only use of  the `MonadCont` instance I've seen in the wild was in the tests!

PR #80, opened 2016-11-15
Possible fix for peekVar

hdgarrood:
I don't understand what's going on enough to be fully confident that this is the right approach, but the test I added does fail with the current code, and passes with this change.

I noticed that in the loop in `_putVar`, in the case where the `consumers` array contains just one consumer where the `peek` property is true, it would go through the loop once, and then repeat the loop without checking the new length of the `consumers` array. This commit essentially expands that loop to include the check of the `consumers` array's length.

### DIFF
--- a/src/Control/Monad/Aff/Internal.js
+++ b/src/Control/Monad/Aff/Internal.js
@@ -46,34 +46,41 @@ exports._peekVar = function (nonCanceler, avar) {
   };
 };
 
+/* jshint loopfunc: true */
 exports._putVar = function (nonCanceler, avar, a) {
   return function (success, error) {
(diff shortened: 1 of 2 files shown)


garyb:
There's something still not quite right here unfortunately... I think some consumers are being run multiple times now, or not being removed or something. Will try and figure out exactly what is going on, and a test case!

natefaubion:
The issue is that the loop does not consider the fact that any synchronous consumers might make additional mutations to the AVar queue. I have a fix for this, just need to write some more test cases.

hdgarrood:
Closing since #83 was merged. :tada:

PR #83, opened 2016-11-28
peekVar fix

### LINKED ISSUE #78: `peekVar` does not work properly
I'm not exactly sure what's happening, but in the new Halogen stuff, when using the new `peekVar`, errors start arising due to an object not having a success property. I'm not sure where it's going wrong as there are some tests in for `peekVar` that show you can still read from a var that has been peeked, etc.

natefaubion:
Fixes #78 

The test runner is a little problematic. With this test, and the previous implementation, it deadlocks and the process exits while still reporting a-ok (the subsequent tests don't run) 😱 . With the fix this test works correctly.

So maybe we should think about some better test machinery...

/cc @hdgarrood

### DIFF
--- a/src/Control/Monad/Aff/Internal.js
+++ b/src/Control/Monad/Aff/Internal.js
@@ -25,7 +25,7 @@ exports._takeVar = function (nonCanceler, avar) {
 
       producer(success, error);
     } else {
-      avar.consumers.push({ success: success, error: error });
+      avar.consumers.push({ peek: false, success: success, error: error });
     }
 
     return nonCanceler;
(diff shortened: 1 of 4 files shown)


natefaubion:
@garyb please confirm that this works for Halogen (I tried it and it worked).

src/Control/Monad/Aff/Internal.js
```diff
@@ -50,28 +50,43 @@ exports._putVar = function (nonCanceler, avar, a) {
...
+        avar.producers.push(function (success, error) {
+          try {
+            success(a);
+          } catch (err) {
+            error(err);
+          }
+        });
+      }
+
+      if (consumers.length) {
```
natefaubion: No need to guard this

garyb:
I've tested it a bit more with Halogen and it does indeed look good to me 👍

natefaubion:
Alright, I've gone ahead and fixed up the weird error issues within Aff I was observing, and added a more resilient test. With this test I've confirmed for `peekVar` that @garyb's implementation resulted in a run time error, @hdgarrood's implementation resulted in a deadlock, and this implementation works correctly.

ping @garyb @jdegoes

garyb:
Looks "good" to me! 👍

jdegoes:
Same here, looks "great" to me! 😉 :shipit:

hdgarrood:
It would be super helpful for me if you could publish a release with this fix in.

garyb:
@hdgarrood just made a release with it in now, `v2.0.2`.

hdgarrood:
Fab, thanks!

PR #85, opened 2016-12-05
Simplify AVar JS

garyb:
Is there a reason not to just enqueue the values like this, rather than having an array of producers? Also dropped the `{}` for unit, as that's unnecessary now.

### DIFF
--- a/src/Control/Monad/Aff/Internal.js
+++ b/src/Control/Monad/Aff/Internal.js
@@ -4,7 +4,7 @@ exports._makeVar = function (nonCanceler) {
   return function (success) {
     success({
       consumers: [],
-      producers: [],
+      queued: [],
       error: undefined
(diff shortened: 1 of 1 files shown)


natefaubion:
I think its so that we could at some point have `putVar` block until the value was taken (to work like MVars, and not like a queue). However, we haven't ever done that, and doing that now may open up a can of worms. @jdegoes?

jdegoes:
Yeah, ultimately I'd like `AVar` to work like `MVar`. It is a change in semantics, may deadlock or otherwise change behavior in code — needs to be thought through.

garyb:
If we were going to do that, I think we'd probably want to introduce an alternative type. We rely upon the current `AVar` semantics in multiple places, and not just in SlamData.

PR #87, opened 2016-12-05
Cancel the other branch when there's succeess in a race

garyb:


natefaubion:
I think this could probably use a test (using `cancelWith`).

garyb:
@natefaubion How's this?

natefaubion:
👍

PR #98, opened 2017-05-08
add tryTakeVar and tryPeekVar

syaiful6:
fixes https://github.com/slamdata/purescript-aff/issues/89

### DIFF
--- a/src/Control/Monad/Aff/Internal.js
+++ b/src/Control/Monad/Aff/Internal.js
@@ -25,6 +25,22 @@ exports._takeVar = function (nonCanceler, avar) {
   };
 };
 
+exports._tryTakeVar = function (nothing, just, nonCanceler, avar) {
+  return function (success, error) {
+    if (avar.error !== undefined) {
+      error(avar.error);
+    } else if (avar.producers.length > 0) {
+      avar.producers.shift()(function (x) {
+        success(just(x));
+        return nonCanceler;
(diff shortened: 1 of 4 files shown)


src/Control/Monad/Aff/Internal.js
```diff
@@ -25,6 +25,22 @@ exports._takeVar = function (nonCanceler, avar) {
...
 };
 
+exports._tryTakeVar = function (nothing, just, nonCanceler, avar) {
+  return function (success, error) {
+    if (avar.error !== undefined) {
+      error(avar.error);
+    } else if (avar.producers.length > 0) {
+      avar.producers.shift()(function (x) {
+        success(just(x));
+        return nonCanceler;
```
natefaubion: I don't think this is necessary. Judging by the other function (which just pass in `success` and `error` to the producer, this should just be equivalent to `success <<< Just`. Which means the JS should be:
```js
avar.producers.shift()(function (x) {
  return success(just(x));
}, error);
```

Same for the `tryPeekVar` case.

syaiful6: updated, i just using the style i see on FFI file for Aff.js, there ```_pure```, ```_throwError``` return nonCanceller. But, i guess you are right, because ```_fmap``` in that file even doesn't return nonCanceller, nor return the return value

PR #99, opened 2017-05-31
Added Discard instance

Risto-Stevcev:


natefaubion:
Thanks for the PR. We no longer use `Canceler` as a return type, but even with `Fiber`, I do not think it's a good idea to add `Discard` instances for stuff like this. It defeats the point of the warning.

PR #102, opened 2017-06-08
Add conversion to ContT

redxaxder:


natefaubion:
Thank you for the PR. I don't think this is a good conversion to make. Aff continuations can only be invoked at most once, which would be unexpected for `ContT`. We purposefully removed the `MonadCont` instance.

PR #103, opened 2017-06-08
Change ErrorT to ExceptT. Include omitted Error type.

redxaxder:


redxaxder:
Oh, I misunderstood what was going on here.

I couldn't find ErrorT in Pursuit so I thought it was a typo for ExceptT, but I now see that transformers used to include ErrorT. 

It was removed here https://github.com/purescript/purescript-transformers/releases/tag/v0.7.1

garyb:
`ExceptT` is the revised `ErrorT`, so this is right :)

natefaubion:
Thank you for the PR! We no longer make this comparison in the README, though, so I'm going to close this.

PR #97, opened 2017-04-06
Delete unlawful MonadPlus instance

rightfold:
Fixes #95.

PR #106, opened 2017-06-29
4.0.0 rewrite

### LINKED ISSUE #18: make avar puts block
If an AVar is full, putting should block until it's empty.

### LINKED ISSUE #91: takeVar/peekVar should have a canceler
`tryVar` and `peekVar` suspend the continuation, but since they use `nonCanceler`, if you try to kill one of these operations, it won't do anything. It should be possible/reasonable to nullify the continuation.

### LINKED ISSUE #94: Rename monadZero instance to monadZeroAff

### LINKED ISSUE #95: Is MonadPlus instance lawful?
`let f = if _ then throwError e else pure unit`

Then `(pure true <|> pure false) >>= f` is not the same as `(pure true >>= f) <|> (pure false >>= f)`. Because in the first example, the exception thrown by `f` is not caught.

### LINKED ISSUE #104: What are the semantics of attempt + cancellation?
If I have this code:

```purescript
do
  canceler <- forkAff do
    res <- attempt do
      delay (Milliseconds 100.0)
      pure unit
    case res of
      Left err -> logShow error
      Right _  -> log "Done"
  delay (Milliseconds 10.0)
  canceled <- cancel canceler (error "Stop!")
  when canceled do
    log "Canceled"
```

What should this log? Should `attempt` mask a cancellation attempt, and prevent cancellation?

natefaubion:
This is a PR to rewrite the internals of `Aff` to be faster and have actual guarantees around cancellation and finalization. There's still a lot to do, but I wanted to go ahead and get some feedback on the implementation.

Closes #97 
Closes #96
Fixes #104 
Fixes #95 
Fixes #94 
Fixes #91 
Fixes #18


natefaubion:
I've added a bunch of tests and worked out a lot of the kinks. I still have more tests to add back pertaining to `ParAff`.

I'd like to detail the behavior around the threading model and exceptions, and the interplay of effect labels. I've removed the `exception` label from `launchAff`, because it is extremely misleading. This is always a lie because it implies you should use `catchException` to remove the label, yet no exception will actually be caught because it is thrown asynchronously. Instead, I guarantee the following: if a thread throws an exception, and there are no other threads observing the exception (via `joinThread`), the exceptions will always be thrown in a fresh stack so as not to interrupt another thread, and still remain observable (eg., by a global `onerror` handler). Alternatively, I've added an `async :: ASYNC` effect to `launchAff` and `runAff`. These are equivalent to starting up a new thread scheduler, so I think it's appropriate for them to carry an appropriate effect.

I've also tried something different here with regards to `liftEff` and `makeAff`:

```purescript
type AffModality eff =
  ( exception ∷ EXCEPTION
  , async ∷ ASYNC
  | eff
  )

-- Needs to have alias inlined because synonym desugaring bug in instance declarations
instance monadEffAff ∷ EffectRowEquals eff1 (exception ∷ EXCEPTION, async ∷ ASYNC | eff2) ⇒ MonadEff eff1 (Aff eff2)

makeAff ∷ ∀ eff a. ((Either Error a → Eff (AffModality eff) Unit) → Eff (AffModality eff) (Canceler eff)) → Aff eff a
```

In other words, any underlying `Eff` actions used to build an `Aff` implicitly carry `EXCEPTION` and `ASYNC` effects. This means that calling `liftEff` will remove the `EXCEPTION` label, as exceptions are an implicit effect in `Aff`. This mostly just works the way you want it to, except in one case.

```purescript
someAff = makeAff \k -> do
  thread <- launchAff do
    delay (Milliseconds 100.0)
    liftEff $ k (Right unit)
  pure $ Canceler (flip killThread thread)
```

Or in other words, take the continuation from one scheduler, launch a new one, and invoke the continuation from the new scheduler. The effect rows won't unify in this case. Even though this is how some things are implemented internally, I'd argue this is an inherently unsafe thing to do, so I've provided `unsafeLaunchAff` and `unsafeLiftEff` functions without the `AffModality` restriction on effect labels. I'm open to removing it, but then we are back to what we currently have.

natefaubion:
@paf31 suggested trying `ASYNC :: # Effect -> Effect` to describe whatever async effects are happening. So `launchAff` would have the following type:

```purescript
launchAff
  :: forall eff1 eff2.
   . Aff eff1
  => Eff (async :: ASYNC (exception :: EXCEPTION | eff1) | eff2) Unit
```

I think this is interesting, and perhaps more accurately describes what's going on.

jdegoes:
@natefaubion Last time I tried that, the type inferencer had major problems, and adding labels of any kind selectively results in user-pain.

What is the pain that adding `ASYNC` solve? The use of string labels to model semantic effects is already provably a failure IMO.

natefaubion:
> @natefaubion Last time I tried that, the type inferencer had major problems, and adding labels of any kind selectively results in user-pain.

Likely before fundeps? The EffectRowEquals constraints uses functional dependencies, so it should be the same as wrapping every Eff with `catchException`. I'd like to use `RowLacks` to actually prevent exception labels from being introduced at all, but the currying of the multiple runtime dictionaries used to encode it has significant overhead.

But you are right, it just may not be worth it, and there are tons of instances in the current ecosystem where `exception` labels are introduce asynchronously and can't be caught.

natefaubion:
I think I will still elide the additional `EXCEPTION` effect from `launchAff`, but will probably remove the other machinery.

syaiful6:
We have ```withResource``` defined on purescript-transformer and I always use it when want similiar feature of Haskell's ```bracket``` function. It look like we implement that here. I wonder what's difference here. Also ```try``` on purescript-transformer have similiar behaviour with ```attempt```. Should we really implement same functionality here?

natefaubion:
@syaiful6 We could probably formulate `Aff` so `catchError` is the primitive operation rather than `attempt`. `attempt` is slightly easier to implement, but you are right, it is otherwise redundant. `bracket` however must be a primitive because it's not as simple as just error handling (which is what `withResource` does). It also affects cancellation behavior. Actions run during acquisition and cleanup _cannot_ be cancelled. If you try to kill a thread during these actions it will block until they have completed.

natefaubion:
This still needs documentation updated, but otherwise the code is ready for review.

natefaubion:
The `ParAff` instance is clearly not stack safe, so I will try to reformulate it with some sort of free applicative.

natefaubion:
SOMEBODY PLEASE REVIEW QUICK (before I'm tempted to do any more to it).

src/Control/Monad/Aff/AVar.purs
```diff
@@ -1,77 +1,81 @@
...
+import Control.Monad.Eff.AVar as AVar
+import Control.Monad.Eff.Class (liftEff)
+import Control.Monad.Eff.Exception (Error)
+import Data.Maybe (Maybe)
+
+toCanceler ∷ ∀ eff. Eff eff Unit → Canceler eff
+toCanceler = Canceler <<< const <<< liftEff
+
+-- | Creates a fresh AVar with an initial value.
+makeVar ∷ ∀ eff a. a → Aff (avar ∷ AVAR | eff) (AVar a)
```
MiracleBlue: This is purely a style opinion, but I'm not a fan of using characters like `→` and `∷` to do pseudo-ligatures, I'd much prefer leaving it up to the editor (and the font in the editor) to do ligatures when supported.  Not a show-stopper, though, just my personal opinion.

src/Control/Monad/Aff/Class.purs
```diff
@@ -13,35 +12,34 @@ import Control.Monad.RWS.Trans (RWST)
...
-  liftAff :: forall a. Aff eff a -> m a
+class MonadEff eff m ⇐ MonadAff eff m | m → eff where
+  liftAff ∷ ∀ a. Aff eff a → m a
 
-instance monadAffAff :: MonadAff e (Aff e) where
+instance monadAffAff ∷ MonadAff e (Aff e) where
   liftAff = id
 
-instance monadAffContT :: MonadAff eff m => MonadAff eff (ContT r m) where
+instance monadAffContT ∷ MonadAff eff m ⇒ MonadAff eff (ContT r m) where
```
MiracleBlue: Same as my previous comment about pseudo-ligatures, although here I think it's particularly tricky due to how small the character is for `⇒` compared to the rest of the text, it makes it much harder to see.

jdegoes:
This looks phenomenal to me. I did not attempt to "bug check" the native code. Going to have to rely on tests for that one. 😆 

The one suggestion I might have is including a few redundant functions in here purely for backward compatible reasons, i.e. making upgrade easier (e.g. `attempt`).

natefaubion:
I wanted to make sure everything worked whether or not we decided to schedule forks (rather than evaluating immediately), and there were some edge cases that failed. That was a red flag to me, so I've refactored the internal Fiber API to be more robust, and at least manually verified that it works and all tests pass regardless of evaluation strategy around forks in case we decide to change it at some point.

natefaubion:
@jdegoes I think we just need to decide whether we want supervised threads to be the default, and what we should do about laws in https://github.com/slamdata/purescript-fork/issues/4#issuecomment-320749770.

jdegoes:
@natefaubion All right. Give me a day to think this over and we can hash out the details tomorrow.

natefaubion:
I've added `attempt`, `apathize`, `cancelWith`, and `nonCanceler`.

natefaubion:
I've changed the default behavior of `forkAff` to be unsupervised, and added a `supervise` combinator instead.

natefaubion:
This is ready to go. Would anyone like to review the README changes?

README.md
```diff
@@ -38,125 +32,157 @@ deleteBlankLines path = do
...
 ```
 
-This looks like ordinary, synchronous, imperative code, but actually operates asynchronously without any callbacks. Error handling is baked in so you only deal with it when you want to.
+This looks like ordinary, synchronous, imperative code, but actually operates
+asynchronously without any callbacks. Error handling is baked in so you only
+deal with it when you want to.
 
-The library contains instances for `Semigroup`, `Monoid`, `Apply`, `Applicative`, `Bind`, `Monad`, `Alt`, `Plus`, `MonadPlus`, `MonadEff`, and `MonadError`. These instances allow you to compose asynchronous code as easily as `Eff`, as well as interop with existing `Eff` code.
+The library contains instances for `Semigroup`, `Monoid`, `Apply`,
+`Applicative`, `Bind`, `Monad`, `Alt`, `Plus`, `MonadEff`, and `MonadError`.
```
garyb: Worth mentioning `Parallel` here too? Or perhaps just "and others" if you'd rather reserve the mention of that for the section below.

README.md
```diff
@@ -38,125 +32,157 @@ deleteBlankLines path = do
...
+well as interop with existing `Eff` code.
 
 ## Escaping Callback Hell
 
-Hopefully, you're using libraries that already use the `Aff` type, so you don't even have to think about callbacks!
+Hopefully, you're using libraries that already use the `Aff` type, so you
+don't even have to think about callbacks!
 
-If you're building your own library, or you have to interact with some native code that expects callbacks, then *purescript-aff* provides a `makeAff` function:
+If you're building your own library, then *purescript-aff* provides a
```
garyb: We usually put library names in backticks rather than making them italic I think. Perhaps we can reword it generally to avoid mentioning the library though, since it seems a bit redundant when you're already reading the readme.

README.md
```diff
@@ -38,125 +32,157 @@ deleteBlankLines path = do
...
+  response <- ajaxGet req
+  log response.body
 ```
 
 ## Eff
 
-All purely synchronous computations (`Eff`) can be lifted to asynchronous computations with `liftEff` defined in `Control.Monad.Eff.Class` (see [here](https://github.com/purescript/purescript-eff)).
+All purely synchronous computations (`Eff`) can be lifted to asynchronous
+computations with `liftEff` defined in `Control.Monad.Eff.Class` (see
+[here](https://github.com/purescript/purescript-eff)).
```
garyb: May as well move the link ref onto the class name (rather than linking with a **here**) while we're doing these updates!

README.md
```diff
@@ -167,66 +193,82 @@ Because Javascript is single-threaded, forking does not actually cause the
...
-   a <- takeVar v
-   liftEff $ log ("Succeeded with " ++ show a)
+example = d
+  v <- makeEmptyVar
+  _ <- forkAff do
+    delay (Milliseconds 50.0)
+    putVar v 1.0
+  a <- takeVar v
+  log ("Succeeded with " <> show a)
 ```
```
garyb: Is it worth briefly mentioning the take/put, etc. sematics `AVar` blocking here? If people aren't familiar with `MVar` it might not be obvious what the difference between this and a `Ref` is, even with the code example.

garyb:
All the new changes to the readme LGTM!

natefaubion:
Edits made, with slightly more explanatory avar stuff.

natefaubion:
@jdegoes @garyb Should we merge this? We can do a pre-release, I guess, since there's quite a bit to look into updating.

jdegoes:
@natefaubion Yes, let's merge and do a pre-release as well.

natefaubion:
🎉

garyb:
🎆 🍾 🎈

PR #107, opened 2017-08-02
Remove erroneous MonadRace references

jkachmar:
It looks like the README hadn't been updated to reflect the changes introduced in purescript/purescript-parallel#10. I've just done a quick pass to remove references to `MonadRace` and `race`, and change     the reference from `par` to `parallel`.


jkachmar:
Very probably obsoleted by the rewrite, closing.

PR #114, opened 2017-08-22
Kill fixes

### LINKED ISSUE #110: What should happen if a canceler throws
For example, with the Alt instance, if a thread succeeds, but in the process of killing the other threads, one of them throws, should the entire Alt throw? Or should we throw the error in a fresh stack and continue with a success?

### LINKED ISSUE #113: Exceptions can't be caught after an interrupt
If a fiber is killed, then exceptions can no longer be caught during finalization. This is because the interrupt is naively taking precedence over all exception handlers.

natefaubion:
Fixes #110 
Fixes #113 

This handles bracket/catch behavior with interrupts and finalizers by tagging each enqueued recovery item with the current interrupt status. When we go to evaluate them, we can check to see if the interrupt status has changed, and decide whether we should filter them out, or run the recovery.

### DIFF
--- a/src/Control/Monad/Aff.js
+++ b/src/Control/Monad/Aff.js
@@ -48,9 +48,9 @@ var Aff = function () {
 
   // Various constructors used in interpretation
   var CONS      = "Cons";      // Cons-list, for stacks
-  var RECOVER   = "Recover";   // Continue with error handler
   var RESUME    = "Resume";    // Continue indiscriminately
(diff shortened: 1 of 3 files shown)


natefaubion:
An alternative would have been to rewrite the recovery stack on interrupt, filtering out catch items, and changing the behavior of enqueued bracket conditions. The implemented solution lets us do it "as we go", so we aren't paying any extra.

PR #117, opened 2017-08-25
Reexport exception functions

natefaubion:


test/Test/Main.purs
```diff
@@ -3,13 +3,13 @@ module Test.Main where
...
 import Control.Alt ((<|>))
-import Control.Monad.Aff (Aff, Canceler(..), runAff_, launchAff, makeAff, try, bracket, generalBracket, delay, forkAff, suspendAff, joinFiber, killFiber, never, supervise)
+import Control.Monad.Aff (Aff, Canceler(..), runAff_, launchAff, makeAff, try, bracket, generalBracket, delay, forkAff, suspendAff, joinFiber, killFiber, never, supervise, Error, EXCEPTION, error, message)
 import Control.Monad.Aff.AVar (AVAR, makeEmptyVar, takeVar, putVar)
 import Control.Monad.Eff (Eff, runPure)
 import Control.Monad.Eff.Class (class MonadEff, liftEff)
 import Control.Monad.Eff.Console (CONSOLE)
 import Control.Monad.Eff.Console as Console
-import Control.Monad.Eff.Exception (Error, EXCEPTION, throwException, error, message)
+import Control.Monad.Eff.Exception (throwException)
```
garyb: Should we re-export this too?

natefaubion: I don't think so. It's an Eff API. You would never use it inside of Aff.

natefaubion:
I removed the reexport of `EXCEPTION` since `Aff` doesn't propagate it anywhere. It only reexports the things relevant to the Aff API.

garyb:
👍

PR #119, opened 2017-08-31
AVar updates

natefaubion:


README.md
```diff
@@ -230,22 +230,23 @@ example = do
 > Got a value: hello
 ```
 
-Likewise, calling `putVar` will queue until it is taken:
+Likewise, calling `putVar` on a filled `AVar` will queue until it is emptied by
+a `takeVar`.
```
garyb: This is a nice and clear way of explaining it, much better than my word salad on Slack.

PR #120, opened 2017-09-02
Compat updates

natefaubion:
I changed the types in Compat to match the FFI examples. It now implicitly adds a `nonCanceler` for canceling the canceler 🤓.


natefaubion:
All good choices. @jdegoes chose `nonInterruptible`, but why not a little drama?

jdegoes:
Something that is invincible cannot be killed, but may die for other reasons (old age?), so that may fit. 😆

PR #135, opened 2017-12-16
ParAff applicative should abort on uncaught exception

### LINKED ISSUE #134: ParAff Apply instance should abort on an unhandled exception
Right now if one side of `apply` throws an exception, it will still wait until the other side completes. We should kill the other side. Since the entire tree is in an error state at that point there's no use in waiting around.

natefaubion:
Fixes #134

### DIFF
--- a/src/Control/Monad/Aff.js
+++ b/src/Control/Monad/Aff.js
@@ -19,7 +19,7 @@ var Aff = function () {
     | Async ((Either Error a -> Eff eff Unit) -> Eff eff (Canceler eff))
     | forall b. Bind (Aff eff b) (b -> Aff eff a)
     | forall b. Bracket (Aff eff b) (BracketConditions eff b) (b -> Aff eff a)
-    | forall b. Fork Boolean (Aff eff b) ?(Thread eff b -> a)
+    | forall b. Fork Boolean (Aff eff b) ?(Fiber eff b -> a)
     | Sequential (ParAff aff a)
 
   */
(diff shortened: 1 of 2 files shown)


jdegoes:
@natefaubion Thanks, will review this today!

test/Test/Main.purs
```diff
@@ -411,6 +411,21 @@ test_parallel = assert "parallel" do
   r2 ← joinFiber f1
   pure (r1 == "foobar" && r2.a == "foo" && r2.b == "bar")
 
+test_parallel_throw ∷ ∀ eff. TestAff eff Unit
```
jdegoes: Worth adding a test here for parallel `a` / `never`.

src/Control/Monad/Aff.js
```diff
@@ -762,26 +762,33 @@ var Aff = function () {
...
-            step    = null;
-            fail    = rhs;
+          // If we have a failure we should kill the other side because we
+          // can't possible yield a result anymore.
+          if (fail) {
             head._3 = fail;
+            tmp     = true;
+            kid     = killId++;
+
+            kills[kid] = kill(early, fail === lhs ? head._2 : head._1, function (/* unused */) {
```
jdegoes: This looks tricky. Don't understand the encoding well enough to say if this is correct.

natefaubion: * `head` = the current Applicative node in the stack
* `head._1` = the left-hand-side node
* `head._2` = the right-hand-side node
* `fail` = the resolved exception
* `lhs` = the resolved value of the left-hand-side parallel operation (`head._1._3`)
* `rhs` = the resolved value of the right-hand-side parallel operation (`head._2._3`)

So, if the left-hand-side resolved to a failure, kill the right-hand-side, otherwise kill the left-hand-side.

natefaubion: I don't think I actually need `rhs` anymore, so I can lose that.

natefaubion:
* I've changed the test to use `never` with a timeout guard around it.

src/Control/Monad/Aff.js
natefaubion: Nevermind, it's used in the success branches.

PR #141, opened 2018-03-23
Patch 1

chexxor:
This question came up in chat, would like to answer it for future users of the lib. Hope I got the code sample right. I'll have to type-check it later, when I can get a PS environment.


thomashoneyman:
@chexxor I'm sorry this was never merged; I'm closing for inactivity.

PR #142, opened 2018-04-02
add Lazy instance for Aff

safareli:
With this instance you can do:

``` purescript
fix \loop -> do
  dostaff
  loop

-- instead of
let loop = do
  dostaff
  loop
```

I also tried to add Lazy instance for ParAff:
``` purescript
instance lazyParAff ∷ Lazy (ParAff eff a) where
  defer f = parallel $ pure unit >>= (f >>> _sequential)
```

but this explodes:
``` purescript
fix \loop -> parallel (delay (Milliseconds 10.0) *> modifyRef ref (add 1)) *> loop
```

If you can think of  simpler test let me know.

### DIFF
--- a/src/Control/Monad/Aff.purs
+++ b/src/Control/Monad/Aff.purs
@@ -36,6 +36,7 @@ import Prelude
 import Control.Alt (class Alt)
 import Control.Alternative (class Alternative)
 import Control.Apply (lift2)
+import Control.Lazy (class Lazy)
 import Control.Monad.Eff (Eff, kind Effect)
 import Control.Monad.Eff.Class (class MonadEff, liftEff)
 import Control.Monad.Eff.Exception (Error, EXCEPTION, error)
@@ -109,6 +110,9 @@ instance monadErrorAff ∷ MonadError Error (Aff eff) where
 
 instance monadEffAff ∷ MonadEff eff (Aff eff) where
   liftEff = _liftEff
+
+instance lazyAff ∷ Lazy (Aff eff a) where
+  defer f = pure unit >>= f
 
 -- | Applicative for running parallel effects. Any `Aff` can be coerced to a
 -- | `ParAff` and back using the `Parallel` class.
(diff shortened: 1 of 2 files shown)


src/Control/Monad/Aff.purs
```diff
@@ -110,6 +111,9 @@ instance monadErrorAff ∷ MonadError Error (Aff eff) where
 instance monadEffAff ∷ MonadEff eff (Aff eff) where
   liftEff = _liftEff
 
+instance lazyAff ∷ Lazy (Aff eff a) where
+  defer f = pure unit >>= f
```
safareli: We can define instance `Lazy (m a)` for any `Monad m`

natefaubion: I don't think so. For example, if you used Maybe and tried to construct a fix point it would probably blow the stack.

natefaubion: It works for any Free-like Monad though.

natefaubion: Where binds are suspended in the data structure, rather.

natefaubion:
> I also tried to add Lazy instance for ParAff:
> but this explodes:

I would hope so! You are spawning an infinite number of Aff threads in parallel! At the very least you'd get OOM.

The only reason it works for Aff is because the recursion is guarded under a bind, which is deferred until it's evaluated by the runtime, but this doesn't work for ParAff since Applicative operators are not lazy themselves.

safareli:
Yeah correct! 💃

safareli:
@natefaubion  can you approve it if you are fine with it?

test/Test/Main.purs
```diff
@@ -641,6 +662,7 @@ main = do
   test_liftEff
 
   void $ launchAff do
+    test_lazy
```
natefaubion: Can we move this test? They are ordered by requirements, so this should go after kill tests.

safareli:
@garyb can you create a new release?

PR #145, opened 2018-04-17
Fix 'bhead is not a function' for forks

### LINKED ISSUE #144: bhead is not a function
The following code crashes with the error "bhead is not a function". I am trying to obtain a fiber which is guaranteed to be cleaned up without having to rely on the supervise function.

```
        fiber <- forkAff do
          bracket
            (forkAff do
              forever do
                delay $ 100.0 # Milliseconds
                traceAnyA "a"
             )
             (killFiber $ error "DONE")
             (\fiber -> do
                delay $ 100.0 # Milliseconds
                joinFiber fiber
             )
        delay $ 50.0 # Milliseconds
        killFiber (error "...") fiber
```

natefaubion:
Fixes #144 

There seems to have been an attempt at a shortcut in evaluation of a `Fork` node. For some reason the state machine jumps immediately to `STEP_BIND` which isn't always valid. It's possible to be within a context where the `Fork` is the last node to evaluate. This appeared in the bug report within a `bracket` because a `Bracket` node uses the `attempt` stack and resets `bhead` and `btail` to null. You could probably trigger this bug with `catchError (..) (forkAff (pure unit))` as well.

### DIFF
--- a/src/Control/Monad/Aff.js
+++ b/src/Control/Monad/Aff.js
@@ -372,7 +372,7 @@ var Aff = function () {
             break;
 
           case FORK:
-            status = STEP_BIND;
+            status = STEP_RESULT;
             tmp    = Fiber(util, supervisor, step._2);
             if (supervisor) {
               supervisor.register(tmp);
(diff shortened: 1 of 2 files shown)


natefaubion:
@felixSchl Do you mind trying this branch out?

PR #147, opened 2018-05-01
0.12 Updates

natefaubion:
Should I move the namespace to `Effect.Aff`?


garyb:
What's happening with the `Aff` version of `AVar` if we're dropping it from here?

natefaubion:
I just thought it would be weird to export `Aff` from `Effect.Async`. I can also just export it under `Aff`, without a prefix.

natefaubion:
I'm going to move Aff bindings to purescript-avar.

PR #149, opened 2018-05-09
Add note about some runtime guaranties

### LINKED ISSUE #148: is it fine if onError, onSuccess is invoked after the computation is canceled?
The lib i'm wrapping has no cancellation, this is what i have now:
```js
exports._sendAsync = function (provider) {
    return function (request) {
        return function(onError, onSuccess) {
            var canceled = false;
            provider.sendAsync(request, function(err, succ) {
                if (canceled) {
                    return;
                } else if (err) {
                    onError(err);
                } else {
                    onSuccess(succ);
                }
            });
            return function (cancelError, onCancelerError, onCancelerSuccess) {
                canceled = true;
                onCancelerSuccess();
            };
        };
    };
};
```

my question is, is the `canceled` staff required, or this version will be fine too:
```js
exports._sendAsync = function (provider) {
    return function (request) {
        return function(onError, onSuccess) {
            provider.sendAsync(request, function(err, succ) {
                if (err) {
                    onError(err);
                } else {
                    onSuccess(succ);
                }
            });
            return function (cancelError, onCancelerError, onCancelerSuccess) {
                onCancelerSuccess();
            };
        };
    };
};
```

safareli:
closes #148

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,9 @@ exports._ajaxGet = function (request) { // accepts a request
       req.cancel(); // cancel the request
       cancelerSuccess(); // invoke the success callback for the canceler
     };
+    // Note:
+    // - If onError/onSuccess is called, then canceler function will never be called,
+    // - If the Aff computation is canceled, called onError/onSuccess will do nothing.
   };
 };
 ```


JordanMartinez:
Looks like this PR's edits either need to happen in another file now or should be added as docs to `Aff` itself.

PR #150, opened 2018-05-12
Remove unnecessary dependencies: free, st, type-equality

zyla:
Since `Aff` no longer uses `Free` under the hood, the dependency is not needed.

I also removed `st` and `type-equality`, because it compiles just fine without them.


garyb:
This was already resolved in 0.12 branch, but thanks :)

Issue #153, opened 2018-07-10
TypeError: Cannot read property 'constructor' of null


felixSchl:
The following code throws `TypeError: Cannot read property 'constructor' of null`. If you remove the `invincible` all works well.

```purescript
module Test.Main where

import Effect.Aff
import Prelude

import Control.Parallel (parSequence_)
import Effect (Effect)
import Effect.Console (log)

main :: Effect Unit
main = do
  launchAff_ do
    parSequence_
      [ do
          invincible do
            delay $ 100.0 # Milliseconds
      , do
          delay $ 50.0 # Milliseconds
          throwError $ error "hi3"
      ]
```

The full stack trace:

```
TypeError: Cannot read property 'constructor' of null
    at Object.isLeft (/home/felix/projects/dn3010/aff-bug-report/output/Effect.Aff/index.js:76:112)
    at join (/home/felix/projects/dn3010/aff-bug-report/output/Effect.Aff/foreign.js:720:16)
    at /home/felix/projects/dn3010/aff-bug-report/output/Effect.Aff/foreign.js:778:19
    at /home/felix/projects/dn3010/aff-bug-report/output/Effect.Aff/foreign.js:669:29
    at runEff (/home/felix/projects/dn3010/aff-bug-report/output/Effect.Aff/foreign.js:80:7)
    at run (/home/felix/projects/dn3010/aff-bug-report/output/Effect.Aff/foreign.js:496:15)
    at /home/felix/projects/dn3010/aff-bug-report/output/Effect.Aff/foreign.js:333:19
    at drain (/home/felix/projects/dn3010/aff-bug-report/output/Effect.Aff/foreign.js:120:9)
    at Object.enqueue (/home/felix/projects/dn3010/aff-bug-report/output/Effect.Aff/foreign.js:141:11)
    at Timeout._onTimeout (/home/felix/projects/dn3010/aff-bug-report/output/Effect.Aff/foreign.js:330:27)
```

PR #154, opened 2018-07-10
Issue/153

felixSchl:
Add a failing and a passing test for #153

### DIFF
--- a/test/Test/Main.purs
+++ b/test/Test/Main.purs
@@ -5,7 +5,7 @@ import Prelude
 import Control.Alt ((<|>))
 import Control.Lazy (fix)
 import Control.Monad.Error.Class (throwError, catchError)
-import Control.Parallel (parallel, sequential, parTraverse_)
+import Control.Parallel (parSequence_, parTraverse_, parallel, sequential)
 import Data.Array as Array
 import Data.Bifunctor (lmap)
 import Data.Either (Either(..), either, isLeft, isRight)
@@ -14,7 +14,7 @@ import Data.Maybe (Maybe(..))
 import Data.Time.Duration (Milliseconds(..))
 import Data.Traversable (traverse)
 import Effect (Effect)
-import Effect.Aff (Aff, Canceler(..), runAff, runAff_, launchAff, makeAff, try, bracket, generalBracket, delay, forkAff, suspendAff, joinFiber, killFiber, never, supervise, Error, error, message)
+import Effect.Aff (Aff, Canceler(..), runAff, runAff_, launchAff, makeAff, try, bracket, generalBracket, delay, forkAff, suspendAff, joinFiber, killFiber, never, supervise, Error, error, message, invincible)
 import Effect.Aff.Compat as AC
 import Effect.Class (class MonadEffect, liftEffect)
 import Effect.Console as Console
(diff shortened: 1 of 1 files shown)


felixSchl:
This PR is not an attempt at fixing this as the Aff internals are still over my head, but hopefully it incentives finding a fix by having CI fail for a legit use-case of the library :)

felixSchl:
I got the tests to pass using this, but it's probably wrong since I do not understand the implications

```
            kills[kid] = kill(early, fail === lhs ? head._2 : head._1, function (/* unused */) {
              return function () {
                delete kills[kid];
                if (tmp) {
                  tmp = false;
                } else if (tail === null) {
                  join(step || fail, null, null); // notice the 'step || fail'
                } else {
                  join(step, tail._1, tail._2);
                }
              };
            });
```

test/Test/Main.purs
```diff
@@ -463,8 +484,8 @@ test_parallel_alt = assert "parallel/alt" do
 test_parallel_alt_throw ∷ Aff Unit
 test_parallel_alt_throw = assert "parallel/alt/throw" do
   r1 ← sequential $
-    parallel (delay (Milliseconds 10.0) *> throwError (error "Nope."))
-    <|> parallel (delay (Milliseconds 11.0) $> "foo")
+        parallel (invincible $ delay (Milliseconds 11.0) $> "foo")
```
natefaubion: What's the reason for adding invincible to this test?

test/Test/Main.purs
```diff
@@ -392,6 +393,15 @@ test_kill_finalizer_bracket = assert "kill/finalizer/bracket" do
...
   eq "completed" <$> readRef ref
 
+test_kill_invincible ∷ Aff Unit
+test_kill_invincible = assert "kill/invincible" do
+  true <$ do
+    f <- forkAff $
+      invincible $
+        delay $ Milliseconds 50.0
+    delay $ Milliseconds 10.0
+    killFiber (error "crash") f
```
natefaubion: I'm not sure exactly what this is testing. It does not verify behavior in any way.

test/Test/Main.purs
```diff
@@ -424,6 +434,17 @@ test_parallel_throw = assert "parallel/throw" $ withTimeout (Milliseconds 100.0)
   r2 ← readRef ref
   pure (isLeft r1 && r2 == "foo")
 
+test_parallel_throw_invincible ∷ Aff Unit
+test_parallel_throw_invincible = assert "parallel/throw/invincible" do
+  map (all identity) $ parSequence
```
natefaubion: `all identity` is `Data.Foldable.and`.

test/Test/Main.purs
```diff
@@ -424,6 +434,17 @@ test_parallel_throw = assert "parallel/throw" $ withTimeout (Milliseconds 100.0)
...
+    [ false <$ do
+        invincible $ delay $ Milliseconds 50.0
+    , false <$ do
+        delay $ Milliseconds 10.0
+        throwError $ error "crash"
+    ] `catchError` \e ->
+      pure $ [ Error.message e == "crash" ]
+
```
natefaubion: Similar to the other test, I'm not sure what this is verifying other than that you've included an `invincible` which presumably triggers a runtime error. I would still like to verify some sort of behavior around `invincible`, since that is valuable in specifying what the behavior should be. So in the case of `parSequence`, I would expect the `invincible` block to fully complete (which I would check by modifying a Ref), but then still observe the exception.

natefaubion:
Note that `invincible` is just built using `bracket`, where the effect is run as part of acquisition (which can't be cancelled), so it's not tied directly to invincible. 

https://github.com/slamdata/purescript-aff/blob/ac1fff841121158b7307f719b69a64e7df625986/src/Effect/Aff.js#L778-L780
The problem here is that we are in a guarded `if (fail)` block, which means `step` is always `null`. We should be passing `fail` into `join`. I think this is a copy-paste error :)

PR #161, opened 2018-08-22
Fix finalization after failure

### LINKED ISSUE #159: Failed pattern match at Effect.Aff
The following code causes an uncaught exception to be thrown:

```purescript
  bracket
    (do
      v <- AVar.empty
      v <$ do
        liftEffect $ launchAff_ do
          delay $ 1.0 # Milliseconds
          AVar.kill (error "boom") v
    )
    (\_ _ ->
      void $ attempt $ pure unit
    )
    \v ->
      forever $ AVar.take v
```

```
/home/felix/projects/aff-bug/output/Effect.Aff/index.js:76
        throw new Error("Failed pattern match at Effect.Aff line 385, column 12 - line 387, column 20: " + [ v.constructor.name ]);
                                                                                                               ^

TypeError: Cannot read property 'constructor' of null
    at Object.isLeft (/home/felix/projects/aff-bug/output/Effect.Aff/index.js:76:112)
    at run (/home/felix/projects/aff-bug/output/Effect.Aff/foreign.js:508:27)
    at /home/felix/projects/aff-bug/output/Effect.Aff/foreign.js:333:19
    at drain (/home/felix/projects/aff-bug/output/Effect.Aff/foreign.js:120:9)
    at Object.enqueue (/home/felix/projects/aff-bug/output/Effect.Aff/foreign.js:141:11)
    at Timeout._onTimeout (/home/felix/projects/aff-bug/output/Effect.Aff/foreign.js:330:27)
    at ontimeout (timers.js:466:11)
    at tryOnTimeout (timers.js:304:5)
    at Timer.listOnTimeout (timers.js:267:5)
```

The error goes away in one of two conditions:

* don't `AVar.take` in the run handler
* don't use `attempt`

natefaubion:
It appears to have only worked coincidentally before. When the
interpreter enters into a finalization block, it puts a `Finalized` node
onto the attempt stack with the current `step`. However if there was
a failure, then step would be null. It "worked" before because there was
stale state, and `fail` could be preserved through finalization. This
bug was exposed by using `catchError` within the finalization block,
which resets the failure state to null. Thus when it resumes after
finalization, both `step` and `fail` are null, which is an invalid
state. The solution is to preserve `fail` in addition to `step` in
a `Finalized` node.

Fixes #159

### DIFF
--- a/src/Effect/Aff.js
+++ b/src/Effect/Aff.js
@@ -460,7 +460,7 @@ var Aff = function () {
             // because it should not be cancelled.
             case RELEASE:
               bracketCount++;
-              attempts = new Aff(CONS, new Aff(FINALIZED, step), attempts, interrupt);
+              attempts = new Aff(CONS, new Aff(FINALIZED, step, fail), attempts, interrupt);
               status   = CONTINUE;
               // It has only been killed if the interrupt status has changed
               // since we enqueued the item.
@@ -484,6 +484,7 @@ var Aff = function () {
               bracketCount--;
               status = RETURN;
               step   = attempt._1;
+              fail   = attempt._2;
               break;
             }
           }
--- a/src/Effect/Aff.js
+++ b/src/Effect/Aff.js
@@ -261,12 +261,11 @@ var Aff = function () {
     // the provided callback in `makeAff` more than once, but it may also be an
     // async effect resuming after the fiber was already cancelled.
     function run(localRunTick) {
-      var tmp, result, attempt, canceler;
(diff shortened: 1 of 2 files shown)


natefaubion:
@felixSchl Can you confirm this fixes your issue?

I have a suspicion that there are other bugs caused by this sort of state falling through, but I can merge this and investigate that in a follow up PR.

felixSchl:
Hey, sorry for the delay. It does indeed appear to fix the problem :+1:

PR #163, opened 2018-10-26
Update README.md

### LINKED ISSUE #158: Mention `try` in the "Dealing with failure" part of the docs
Surprisingly there's nothing about it at all.

Related: https://github.com/slamdata/purescript-affjax/issues/128

akheron:
Add a note about using `ExceptT` instead of throwing `Error`s in the `Aff` context.

Add an example that uses `try`.

Fixes #158


garyb:
Thanks!

PR #165, opened 2018-12-14
add fiberCanceler

safareli:
It's pretty useful when you are doing `launchAff` and don't want to expose `Fiber a`:
```
foo :: Effect Canceler
foo = fiberCanceler <$> launchAff never
```


natefaubion:
Thanks!

PR #171, opened 2019-03-23
Fix interrupt bugs wrt bracket masking

### LINKED ISSUE #170: Bracket aquire function gets killed
I am experiencing some very odd behavior, to say the least. Unfortunately, I am of yet unable to reproduce this issue in a small test case, but I am still working on it so I'll have to just report the symptoms for the time being.

We are observing the acquire part of a bracket getting killed part way through it *and* the clean up handler running. Because the acquire handler had not finished running, the clean up handler gets an "undefined" value passed in, causing all sorts of trouble.

Hopefully I can recreate this in a small test case, but the outlook is rather grim.

**EDIT**: Using the code from #169 the error goes away, but the fiber never gets killed at all.

natefaubion:
Currently when we check interrupt status, we are not considering the
bracket masking status. When we enter a masked state (`bracketCount > 0`),
it should be impossible for an interrupt to terminate evaluation, but
without the bracket check in CATCH, RESUME, and RELEASE, this is
violated.

*   In a nested async generalBracket, the wrong handler is enqueued.
Since we are in a masked state, it should be impossible for the `killed`
branch to be invoked. However, if an interrupt occurs it is currently
_always_ invoking `killed` when the branch is actually `completed`.
*   In a nested bracket, an interrupt will terminate evaluation of the
inner bracket even though we are in a masked state. This can also happen
with an inner `catch` in a masked state.

We need to always consider the masked state (`bracketCount > 0`) when we
discriminate the interrupt state.

Fixes #170

### DIFF
--- a/src/Effect/Aff.js
+++ b/src/Effect/Aff.js
@@ -411,14 +411,14 @@ var Aff = function () {
             attempts = attempts._2;
 
             switch (attempt.tag) {
-            // We cannot recover from an interrupt. Otherwise we should
+            // We cannot recover from an unmasked interrupt. Otherwise we should
             // continue stepping, or run the exception handler if an exception
             // was raised.
             case CATCH:
               // We should compare the interrupt status as well because we
               // only want it to apply if there has been an interrupt since
               // enqueuing the catch.
-              if (interrupt && interrupt !== tmp) {
+              if (interrupt && interrupt !== tmp && bracketCount === 0) {
                 status = RETURN;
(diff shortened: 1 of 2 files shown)


felixSchl:
For what it's worth we applied this change to a fairly large codebase without having witnessed any negative side effects so far, and the problem we saw is fixed. We'll be running with these changes from now on and report if anything comes up. Thank you for looking into this and fixing it!

Issue #172, opened 2019-05-01
Exceptions in pure code


ford-prefect:
Currently, if we have bad code that throws an exception despite being annotated as being pure, it seems to put `Aff` into a bad state where no more callbacks occur. I've written up an example at:

    https://github.com/ford-prefect/purescript-sandbox

This is of course just bad code, but it _might_ make sense to be defensive against this if there isn't a significant cost, as there are cases with a significant surface of JS interop, which can make for more common programming errors.

ford-prefect:
@natefaubion you mentioned performance concerns w.r.t. adding try/catches to deal with such programming errors. Is this something you've quantified (or that we could try to)? Might help make a more informed decision about the cost.

natefaubion:
There are some basic benchmarks in the repo for pure binds. I would run it as is, apply a try/catch around the `STEP_BIND` case in the interpreter, and then see what the difference is. If there is a significant performance degradation, then I would try lifting the try/catch out of the main loop like we do for `runSync` and benchmark that.

ford-prefect:
Thanks @natefaubion -- I don't see a perf impact in the benchmark. `pulp test` takes ~4.8s to run with and without that change. I've made a PR for this (feel free to modify or ignore if there is a better way to do it, or I can do that too if you like).

reactormonk:
This bug just cost me two days, so I wouldn't mind it being merged.

ford-prefect:
Closed by #173

PR #173, opened 2019-05-13
Try to recover from effectful code that is annotated as pure

ford-prefect:
This puts a try-catch around the bind step in order to try to catch code
that is not expected to have effects but does (by throwing).

While this is clearly a programming error, it is certainly not
implausible that effects creep into FFI code inadvertently for Aff
users. This allows us to be defensive in that case and recover, rather
than have all asynchronous options quitely not work after.

The addition of the try-catch does not seem to have a performance impact
(tested on Node v10.15.3). With and without this change, `pulp test`
takes ~4.8s to run.

There is a discussion around this on #172.

### DIFF
--- a/src/Effect/Aff.js
+++ b/src/Effect/Aff.js
@@ -274,12 +274,18 @@ var Aff = function () {
         switch (status) {
         case STEP_BIND:
           status = CONTINUE;
-          step   = bhead(step);
-          if (btail === null) {
-            bhead = null;
-          } else {
-            bhead = btail._1;
-            btail = btail._2;
+          try {
+            step   = bhead(step);
+            if (btail === null) {
+              bhead = null;
+            } else {
+              bhead = btail._1;
+              btail = btail._2;
+            }
+          } catch (e) {
+            status = RETURN;
+            fail   = util.left(e);
+            step   = null;
           }
           break;
 


reactormonk:
Would it be possible to review this PR?

natefaubion:
The only concern I had with this is whether or not we want to use the interrupt channel instead of the error channel, since I'd consider an exception in "pure" code something you would not want to recover from _in Aff_.

ford-prefect:
I can't say I'm super confident about my understanding of the code, but modifying the bind step like so seems to be what you're suggesting?

```purescript
          case STEP_BIND:
            status = CONTINUE;
            try {
              step   = bhead(step);
              if (btail === null) {
                bhead = null;
              } else {
                bhead = btail._1;
                btail = btail._2;
              }
            } catch (e) {
              status    = COMPLETED;
              interrupt = util.left(e);   
              step      = interrupt;
            }
            break;
```

natefaubion:
Yes, that seems correct.

https://github.com/purescript-contrib/purescript-aff/blob/390857f9143e1a52f7403d05b14c9ca79d356737/src/Effect/Aff.js#L575-L577

ford-prefect:
I've updated the PR to reflect this now.

natefaubion:
Actually, on second thought as I look at this, there's a can of worms here that I am not prepared to address right now. Your original fix was simple and obvious enough. Sorry for the backtracking.

ford-prefect:
No worries! `git reflog` ftw :-)

natefaubion:
Thanks!

PR #177, opened 2019-08-03
General errors

eric-corumdigital:
Includes #169.

I expect to receive a lot of backlash over breaking backwards compatibility. What I hope is that this can start discussion about how to move forward.


src/Effect/Aff.purs
```diff
@@ -101,185 +103,195 @@ instance monadRecAff ∷ MonadRec Aff where
...
--- | into an `Either`.
-attempt ∷ ∀ a. Aff a → Aff (Either Error a)
-attempt = try
+-- | A version of `catchError` that can map the error type.
+catch ∷ ∀ e1 e2 a. Aff e1 a → (e1 → Aff e2 a) → Aff e2 a
+catch = _catchError
+
+-- | A monomorphic version of `try` that can map the error type. Catches thrown
+-- | errors and lifts them into an `Either`.
+attempt ∷ ∀ e1 e2 a. Aff e1 a → Aff e2 (Either e1 a)
```
safareli: There should be `Void` instead of `e2` i suppose

eric-corumdigital: `Aff e a` does not mean an error of type `e` is _definitely_ thrown; it means an error of type `e` _may_ be thrown.

safareli: Yes but if you do `attempt` on some Aff `computation` it must eliminate all possible errors, which means resulting Aff must throw nothing. If you stick `Void` there it would mean `Void may be throw` but as there is no value of type Void it means nothing can be thrown.

garyb: That would prevent you from `attempt`ing inside an exiting `Aff e1`, unless you `lmap absurd (attempt ...)` though, which would be a bit annoying.

safareli: good point

eric-corumdigital:
By the time I achieve what I want it is too far incompatible with the original Aff. I am instead working on a new library that builds from the lessons of Aff. I will retract both these PRs.

PR #180, opened 2019-09-23
simplify definition of never and nonCanceler

safareli:
`_ -> pure mempty` ~~ `_ -> mempty` ~~ `mempty`


thomashoneyman:
You're correct that these are all equivalent, but I find it a little easier to see what's going on when everything is written out -- `mempty` can be a bit magical as it can fit the shape of so many different things. As this didn't get a response from other maintainers I'm going to go ahead and close it, but thank you for the PR!

PR #186, opened 2019-12-29
feat: export Fiber constructor

srghma:


natefaubion:
I would prefer not to do this. I consider it a design mistake that Fiber isn't a completely opaque foreign type. Why do need it exported? There isn't anything you can do with the constructor that you can't do with the public API.

srghma:
ok, understand, tnx

Issue #61, opened 2016-06-06
Imports for the readme examples?


chris-martin:
I'd love to see the imports to get the documentation examples to compile. In particular I'm curious where I can find the `loadFile` function.

jdegoes:
Unfortunately, those are all made up, although there might be some nodejs wrappers for file system operations (there are certainly `Aff` wrappers for databases like MongoDB).

The best place to get working examples right now is [here](https://github.com/slamdata/purescript-aff/blob/master/test/Test/Main.purs).

chris-martin:
Ah, gotcha.

FWIW `purescript-node-fs-aff` provides an `Aff`-based filesystem operations.

therealnicksaunders:
I would really like to see more type annotations myself...

PR #188, opened 2020-02-26
Add disclaimer about non-working examples in readme

milesfrain:
Related to #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ An asynchronous effect monad and threading model for PureScript.
 
 # Example
 
+Note that these documentation examples are partially pseudocode and are unlikely to compile. Working examples can be found [here](https://github.com/purescript-contrib/purescript-aff/blob/master/test/Test/Main.purs).
+
 ```purescript
 main = launchAff do
   response <- Ajax.get "http://foo.bar"


milesfrain:
We could alternatively update the examples so they compile, but that becomes noisy and shifts focus towards how to use the other libraries. For example, this is what's required to get the first example working.

``` purs
import Affjax as Ajax
import Affjax.ResponseFormat as ResponseFormat
import Data.Either (Either(..))
import Effect (Effect)
import Effect.Aff (Fiber, launchAff)
import Effect.Class.Console (log)

main :: Effect (Fiber Unit)
main = launchAff do
  result <- Ajax.get ResponseFormat.string "http://example.com"
  case result of
    Left err -> log $ "GET failure: " <> Ajax.printError err
    Right response -> log $ "GET response: " <> response.body
```

Another option is to just show examples involving `delay` (as is done in the tests).
``` purs
import Effect (Effect)
import Effect.Aff (Fiber, Milliseconds(..), delay, launchAff)
import Effect.Class.Console (log)

main :: Effect (Fiber Unit)
main = launchAff do
  log "hello"
  delay $ Milliseconds 1000.0
  log "world
```

PR #190, opened 2020-04-26
Use Ubuntu Bionic for Travis CI

peterbecich:
Update from Ubuntu 14.04 to 18.04

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
-dist: trusty
+dist: bionic
 sudo: required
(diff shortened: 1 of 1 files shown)


thomashoneyman:
Thanks for updating this, @peterbecich! However, the Contributors libraries are all moving to GitHub Actions (via the [setup-purescript](https://github.com/purescript-contrib/setup-purescript) action), and so we're going to be moving on from Travis. That update should affect this library sometime in the next month or so.

PR #193, opened 2020-09-05
Update README.md with correct import of `log`

osfameron:
Clarify where to get `log` from, as it's non-obvious (and the linked test-case doesn't provide an example of it.)


thomashoneyman:
At the moment the entire README is written without imports -- even some of the functions are made up, like `loadFile`, to illustrate the point without necessarily being runnable code. #61 is a separate discussion of this issue.

We're in the midst of polishing up all the Contributors libraries, including the README files and documentation, and we'll keep in mind that several folks have found the lack of imports confusing. However, I'm going to close this just because the rest of the README will still need fixing and we're about to do an overhaul of the content anyway.

Still, thank you for the PR!

PR #194, opened 2020-09-12
Add roles declarations to allow safe coercions

kl0tl:
This allows terms of type `Aff a` and `ParAff a` to be coerced to type `Aff b` and `ParAff b` when `Coercible a b` holds, hence allowing the zero cost `coerce` instead of `map wrap` and `map unwrap` to introduce and eliminate newtypes under asynchronous computations for instance.


thomashoneyman (approved):

JordanMartinez:
@kl0tl Could you rebase this branch on top of `main`? That should fix CI. I tried to do that myself, but got rejected when I tried to push the changes to your branch.

kl0tl:
Thank you for rebasing for me Thomas!

PR #203, opened 2021-06-04
Make launchAff_ only work on `Aff Unit`

### LINKED ISSUE #202: Make `launchAff_` operate only on `Aff Unit`
**Problem**

I’m finding launchAff_ to be quite dangerous. When I have an interface that requires an Effect Unit but I need to perform some network calls I usually provide the callback with launchAff_ . Now I ran into a situation where I used launchAff_ on an Aff (Either Error Unit) which is not what I intended. Do you think it would hurt much if launchAff_ only worked on Aff Unit values? That way it would be similar to the warning about implicitly discarding values in a do block.

**Describe the solution you'd like**

`launchAff_` has the signature
```purescript
launchAff_ ∷ Aff Unit → Effect Unit
```

**Additional context**
I'm aware that this should cause a major version bump since it's a breaking change. I'm hopeful however, that *when* this change requires somebody to change their code it will improve its clarity or even catch potential bugs.

i-am-the-slime:
**Description of the change**

 1. Adds transitive dependencies to spago.dhall so the new version of Spago builds
 2. Closes #202 
---

**Checklist:**

- [x] Added the change to the changelog's "Unreleased" section with a link to this PR and your username
- [x] Linked any existing issues or proposals that this pull request should close
- [x] Updated or added relevant documentation in the README and/or documentation directory
- [x] (not applicable) ~Added a test for the contribution~


thomashoneyman (approved):

JordanMartinez:
CI fails because of some unused name warnings

i-am-the-slime:
Maybe it works like this?

JordanMartinez:
I've restarted CI.

And just to clarify, we're not expecting to merge this until the next round of breaking changes in the ecosystem, correct?

thomashoneyman:
Yes, I'd like to leave this open until we're ready for a new breaking release of `aff`. That might come at the next compiler release, but it might come sooner if we have other breaking changes we can batch together.

sigma-andex:
This could go in to the 0.15 release as well, right?

PR #206, opened 2021-11-06
Update README.md

jamesdbrock:
https://discourse.purescript.org/t/how-does-an-aff-monad-work/2637/4

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,10 @@
 
 An asynchronous effect monad and threading model for PureScript.
 
+`aff` is in the same space as JavaScript Promises, but with additional support for cancellation (async exceptions) and resource management (`bracket`). It’s cooperative multi-tasking which yields on async constructors (`makeAff`).
+
+For OS-thread parallelism in Node.js, see [node-workerbees](https://github.com/natefaubion/purescript-node-workerbees).
+
 ## Installation
 
 Install `aff` with [Spago](https://github.com/purescript/spago):


README.md
```diff
@@ -8,6 +8,10 @@
 
 An asynchronous effect monad and threading model for PureScript.
 
+`aff` is in the same space as JavaScript Promises, but with additional support for cancellation (async exceptions) and resource management (`bracket`). It’s cooperative multi-tasking which yields on async constructors (`makeAff`).
+
+For OS-thread parallelism in Node.js, see [node-workerbees](https://github.com/natefaubion/purescript-node-workerbees).
```
thomashoneyman, line 13: Since node-workerbees is experimental, I’m not sure we should link it here. What do you think?

README.md
```diff
@@ -8,6 +8,10 @@
 
 An asynchronous effect monad and threading model for PureScript.
 
+`aff` is in the same space as JavaScript Promises, but with additional support for cancellation (async exceptions) and resource management (`bracket`). It’s cooperative multi-tasking which yields on async constructors (`makeAff`).
```
thomashoneyman, line 11: I find this a little easier to parse, but no need to apply the suggestion if you feel differently!

```suggestion
`aff` is in the same space as JavaScript Promises, but with additional support for cancellation (async exceptions) and resource management (`bracket`). It provides cooperative multi-tasking which yields on async constructors (`makeAff`).
```

README.md
jamesdbrock: I don't have a strong opinion about that.

jamesdbrock: Maybe a __Related Packages__ section further down in the README?

jamesdbrock: We should probably put this in related packages too: https://pursuit.purescript.org/packages/purescript-avar/4.0.0

